PR #27, opened 2016-05-12
Add global accessor using the get/set concept

sagebind:
Reference code for issue #28.


AndrewCarterUK:
Like the idea, given that in PHP7 there is a difference between `Error` and `Exception`, I think it would it make sense to name them `Exception` (otherwise people might think they extend `Error`).

sagebind:
@AndrewCarterUK Ah, yes of course. Leftover from when those classes _did_ extend `Error`...

src/Loop.php
```diff
@@ -40,13 +40,77 @@ public static function execute(callable $callback, LoopDriver $driver)
...
+     * @return bool Indicates if an event loop was removed.
+     */
+    public static function drop()
+    {
+        if (null === self::$driver) {
+            return false;
+        }
+
+        // Stop the event loop first.
+        self::stop();
```
kelunik, line 94: This might not work. At least in Amp, `stop()` currently doesn't tick the event loop, it just makes it not entering the next loop iteration, see https://github.com/amphp/amp/blob/master/lib/NativeReactor.php#L130-L142

src/Loop.php
```diff
@@ -40,13 +40,77 @@ public static function execute(callable $callback, LoopDriver $driver)
...
+        // Remove the object reference.
+        self::$driver = null;
```
kelunik, line 96: I don't think this will work the way it is.

I think that's the thing you want to do:

``` php
Loop::execute(function () {
    Loop::drop();
    Loop::execute(function () { /* ... */ });
});
```

In that case, the old loop is still referenced and can't be garbage collected. `run` will never return and `finally` never be executed here: https://github.com/coderstephen/event-loop/blob/5e9e04d920085e624d5c91ebcb5ef99fcf407185/src/Loop.php#L20-L33

sagebind, line 96: Ah, I see what you mean and you're right. Perhaps `$previousDriver` should be a static property stack, which `drop()` can discard. e.g.

``` php
public static function execute($callable, $driver) {
    if (!self::$driverStack) self::$driverStack = new SplStack();
    self::$driverStack->push(self::$driver);
    // ...
}
```

Then unset the stack in drop. 

Bear with me, I'm on mobile 😵

kelunik, line 96: That has still the same issue if I'm right, because `run` will never return.

bwoebi, line 96: Correct, you have to break out, else it will be still referenced from inside the stack in the event loops run function.

kelunik:
Closing as it doesn't work as the loop will always be present in the backtrace.

PR #30, opened 2016-05-13
Add LoopDriverFactory

kelunik:
The LoopDriverFactory SHOULD be used by driver implementations to set a default event loop driver.

This can be achieved by having a Composer autoloader like that:

``` json
{
    "autoload": {
        "files": ["src/bootstrap.php"]
    }
}
```

Where `src/bootstrap.php` contains the following code:

``` php
use Interop\Async\EventLoop\LoopDriverFactory;

class MyDriverFactory implements LoopDriverFactory
{
    public function create()
    {
        return new MyDriver();
    }
}

Loop::setFactory(new MyDriverFactory());
```

This ensures the user doesn't have to care about setting the actual driver. A user just has to require the specific event loop driver package he / she wants to use.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -54,7 +54,7 @@ final class Loop
     /**
      * Create a new driver if a factory is present, otherwise throw.
      *
-     * @throws \LogicException if no factory is set
+     * @throws \LogicException if no factory is set or no driver returned from factory
      */
     private static function createDriver()
     {
@@ -62,7 +62,13 @@ final class Loop
             throw new \LogicException("Can't create an event loop driver without a factory.");
         }
 
-        return self::$factory->create();
+        $driver = self::$factory->create();
+
+        if (!$driver instanceof LoopDriver) {
+            throw new \LogicException("LoopDriverFactory didn't return a LoopDriver.");
+        }
+
+        return $driver;
     }
 
     /**
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -18,10 +18,12 @@ final class Loop
 
     /**
      * Set the factory to be used to create a driver if none is passed to
-     * self::execute.
+     * self::execute. A default driver will be created if none exists yet
+     * to support synchronous waits in traditional applications.
      */
     public static function setFactory(LoopDriverFactory $factory = null) {
         self::$factory = $factory;
+        self::$driver = self::$driver ?: self::createDriver();
     }
 
     /**
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -7,9 +7,22 @@ final class Loop
     use Registry;
 
     /**
+     * @var LoopDriverFactory
+     */
+    private static $factory = null;
+
+    /**
      * @var LoopDriver
      */
     private static $driver = null;
+
+    /**
+     * Set the factory to be used to create a driver if none is passed to
+     * self::execute.
+     */
+    public static function setFactory(LoopDriverFactory $factory = null) {
+        self::$factory = $factory;
+    }
 
     /**
      * Execute a callback within the scope of an event loop driver.
@@ -19,11 +32,12 @@ final class Loop
      *
      * @return void
      */
-    public static function execute(callable $callback, LoopDriver $driver)
+    public static function execute(callable $callback, LoopDriver $driver = null)
     {
-        $previousDriver = self::$driver;
+        $driver = $driver ?: $this->createDriver();
         $previousRegistry = self::$registry;
 
+        $previousDriver = self::$driver;
         self::$driver = $driver;
         self::$registry = [];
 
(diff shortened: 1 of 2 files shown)


kelunik:
It can be discussed whether `LoopDriverFactory` should be an interface or a simple `callable`, but I think it's fine having an interface.

src/Loop.php
```diff
@@ -5,22 +5,36 @@
...
      * @param callable $callback The callback to execute
      * @param LoopDriver $driver The event loop driver
      *
      * @return void
      */
-    public static function execute(callable $callback, LoopDriver $driver)
+    public static function execute(callable $callback, LoopDriver $driver = null)
     {
-        $previousDriver = self::$driver;
+        $driver = $driver ?: $this->createDriver();
```
trowski: Should be `self::createDriver()` 😸

trowski:
This looks great. Libraries can set their own factories as the default, but users can always override if they wish.

src/Loop.php
kelunik: :see_no_evil:

bwoebi:
I think using a simple callable is really good enough here; especially when an user wants to use his own factory, it ends up being more usable.

src/Loop.php
```diff
@@ -33,6 +47,20 @@ public static function execute(callable $callback, LoopDriver $driver)
...
+     *
+     * @throws \LogicException if no factory is set
+     */
+    private static function createDriver()
+    {
+        if (self::$factory === null) {
+            throw new \LogicException("Can't create an event loop driver without a factory.");
+        }
+
+        return self::$factory->create();
```
assertchris: This should validate that the created driver is an instance of `LoopDriver`, else there's not much point to people being able to define their own.

kelunik: Totally agree, but how does it relate to:

> else there's not much point to people being able to define their own.

trowski: This is where PHP 7 return types would be handy. 😄

I suppose we could emulate that behavior and check for a `LoopDriver` instance and throw if something else is returned.

kelunik: @trowski Exactly. Already did that.

trowski: Great! The notification had me looking at an outdated diff apparently.

kelunik: The commit might have been just a minute afterwards.

assertchris: @kelunik 

> Totally agree, but how does it relate to...

If someone can create a factory that returns a string, other internal code will break horribly. We're in agreement that this is a good idea (from the sounds of things it's already implemented).

kelunik: I understand that it breaks, that's why I committed a fix. I just didn't understand how it's related to _"people being able to define their own"_.

assertchris: The line I linked to is a method that gets called on a user-defined object. If the user-defined object returns an invalid value for `create()` then it can break other internal code that depends on `create()` returning an implementation of `LoopDriver`. That is how it relates to "people being able to define their own".

kelunik:
Can we merge it or should I change it to be a simple `callable` if we anyway can't use return types?

AndrewCarterUK:
I'd try and avoid discussion on anything more than implementation details on a PR.

If there's a significant design choice to be had, we should open an issue, summarise any arguments made so far and then start a discussion.

kelunik:
@AndrewCarterUK That's an implementation detail.

AndrewCarterUK:
Okay, I personally prefer the interface to a callable because it's more explicit (I am a factory that does this).

However, if other people disagree I'd prefer we discuss in an issue until we reach consensus.

kelunik:
@bwoebi Could you please open an issue if you want to change it to a callable or respond that we can merge it?

bwoebi:
@kelunik opened #42 to discuss this.

kelunik:
Should we go with `make` or `create`?

bwoebi:
CreateDriver is fine.

kelunik:
You mean I should rename it from `create` to `createDriver`?

bwoebi:
Oh, in factory. I thought you were talking about the static function on loop.

Nah, there just `create` is fine.

kelunik:
I just applied the changes required for the registry (we also need a default one) and added the running bool, to reset loop / factory if a new factory is set and no loop is running.

I think this should be ready to merge then.

src/Loop.php
```diff
@@ -67,7 +80,8 @@ private static function createDriver()
         $driver = self::$factory->create();
 
         if (!$driver instanceof LoopDriver) {
-            throw new \LogicException("LoopDriverFactory didn't return a LoopDriver.");
+            $type = is_object($driver) ? "an instance of " . get_class($driver) : gettype($driver);
+            throw new \LogicException("Factory returned {$type}, but must return an instance of LoopDriver.");
```
bwoebi: s/Factory/LoopDriverFactory/

kelunik: As there's no other factory, I think that's fine. But we could create another exception for it.

bwoebi: Nah, no separate exception, I'd just love to be more explicit in error message.

kelunik: Done.

bwoebi:
Looks good to me now ... I'd like to merge it soon, in case nobody objects until tomorrow.

trowski:
LGTM. Let's rename it to `Interop\Async\Loop\DriverFactory` once #40 is merged.

PR #31, opened 2016-05-13
Add loop bound state registry

kelunik:
The registry allows to store information bound to a specific loop. This can be useful to store state of globals like a DNS resolver.

**Rationale for a global DNS resolver**

Nobody wants to inject a resolver into everything that does some sort of I/O. Using IP addresses everywhere instead of hostnames also doesn't make sense.

Therefore there should be a global default DNS resolver. But this resolver needs a read watcher for the response. If the loop gets swapped and a DNS request is made, the resolver would never get the read event, since the stream is only watched in the old loop. The resolver also doesn't get any info about the loop swap, so it can't use a new watcher.

With a loop bound registry, it can just store the connection in the loop and will use a new connection when the loop is swapped, because the state doesn't exist any longer. Once the loop is put back, it can use the old connection again.

This API should only be used by real global objects. Most packages won't need to use this API, but it's required for the mentioned use cases.

### DIFF
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Interop\Async\EventLoop;
+namespace Interop\Async;
 
 trait Registry {
     /**
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -2,7 +2,8 @@
 
 namespace Interop\Async;
 
-trait Registry {
+trait Registry
+{
     /**
      * @var array
      */
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Interop\Async\EventLoop;
+
+trait Registry {
+    /**
+     * @var array
+     */
+    private static $registry = null;
+
+    /**
+     * Stores information in the loop bound registry. This can be used to store
+     * loop bound information. Stored information is package private.
+     * Packages MUST NOT retrieve the stored state of other packages.
+     *
+     * Therefore packages SHOULD use the following prefix to keys:
+     * `vendor.package.`
+     *
+     * @param string $key namespaced storage key
+     * @param mixed $value the value to be stored
+     *
+     * @return void
+     */
+    public static function storeState($key, $value)
+    {
+        if (self::$registry === null) {
+            throw new \RuntimeException('Not within the scope of an event loop driver');
+        }
+
+        if ($value === null) {
+            unset(self::$registry[$key]);
+        } else {
+            self::$registry[$key] = $value;
+        }
+    }
+
+    /**
+     * Fetches information stored bound to the loop. Stored information is
+     * package private. Packages MUST NOT retrieve the stored state of
+     * other packages.
+     *
+     * Therefore packages SHOULD use the following prefix to keys:
+     * `vendor.package.`
+     *
+     * @param string $key namespaced storage key
+     *
+     * @return mixed previously stored value or null if it doesn't exist
+     */
+    public static function fetchState($key)
+    {
+        if (self::$registry === null) {
+            throw new \RuntimeException('Not within the scope of an event loop driver');
+        }
+
+        return isset(self::$registry[$key]) ? self::$registry[$key] : null;
+    }
+}
--- a/test/RegistryTest.php
+++ b/test/RegistryTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Interop\Async\EventLoop;
+
+class RegistryTest extends \PHPUnit_Framework_TestCase {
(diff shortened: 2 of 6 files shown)


bwoebi:
Looks good to me; I'd like to have someone else merge it though.

kelunik:
I don't think we have consensus here yet, or do we?

kelunik:
Not sure whether `storeState` / `putState` / `setState`. I don't like `setState`, it makes me think it's an event loop state, not something not actually related to the loop driver but just data.

bwoebi:
@kelunik yeah, I'm not sure on consensus, that's why I'd like to have someone else merging it when it's clear that there's consensus … as I'm biased as it was my idea ;-)

bwoebi:
@kelunik either we have get/set or store/fetch. Both variations are fine with me (but not set/fetch).

assertchris:
What about putting the registry stuff into a trait, or a subclass? Subclass may be slower than desired. Is anything else likely to need this registry stuff?

bwoebi:
Having a trait to do that... hmm... (You mean additionally to the interface methods, right?)

assertchris:
@bwoebi yeah, in addition.

kelunik:
What's the benefit of moving it out when just the `Loop` uses it?

assertchris:
It's additional responsibility, that could be tested and documented on its own.

kelunik:
The registry has to be swapped with the loop, how does that work with a trait?

assertchris:
Traits are literally compiler-assisted copy and paste. It would work in exactly the same way as if the code was inside the loop class.

kelunik:
In that case the trait would rely on a variable being present in the loop class. I don't see the benefit here. Just a disadvantage that we rely on a static variable being present in the trait.

Separation of concerns is fine, but not for the sake of it if you have to use a trait for it. Because that doesn't separate anything, it just moves it out, so it can be used in multiple places.

assertchris:
If the property is on the trait then the class would rely on the variable being present. That's an ok relationship since the class already knows it's using the trait. Alternatively, there are abstract method signatures to ensure traits have access to the data from the classes that use them.

They're separate responsibilities. I won't bore you with the benefits of SRP (there are good books which do a much better job of explaining it than I do), but essentially one trades speed for isolated testing, documentation, maintenance and limited side effects to modifying behaviour.

kelunik:
Yes, but this is a utility class that just has static methods and only `execute` changes state at all, so the number of site effects is quite limited.

> If the property is on the trait then the class would rely on the variable being present.

Traits can't define static members in PHP, otherwise I'd probably be fine with it.

bwoebi:
@kelunik Why should traits not be able to define static members? Works in 5.5 all the way up to current master.

@assertchris Thinking about it, I don't think traits are a very good way to enforce SRP. Please leave it in. I'd like to have one class with all the public methods viewable at once. A trait only hides these.
In general, it _is_ fine with SRP. The responsibility is managing the loop, and part of it is managing the attached state too.

The PR is LGTM to me and shall be merged.

kelunik:
@bwoebi Seems like the German translation is just plain wrong then.

> Innerhalb von Traits kann auf statische Variablen zugegriffen werden, aber diese können nicht in Traits definiert werden. Statische Methoden können hingegen auch durch Traits für die darstellende Klasse definiert werden. http://php.net/manual/de/language.oop5.traits.php

bwoebi:
@kelunik well… Fixed in doc commit 339170.

kelunik:
@assertchris I think it's fine as trait then. Refactored it and added tests.

It ensures we can test the registry without an event loop / outside of `Loop::execute`.

src/Registry.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Interop\Async;
+
+trait Registry {
```
AndrewCarterUK: Opening brace should be on a new line according to lord FIG.

test/RegistryTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace Interop\Async;
+
+class RegistryTest extends \PHPUnit_Framework_TestCase {
```
AndrewCarterUK: Opening brace should be on a new line according to lord FIG (as with all the methods here).

kelunik: Done.

src/Registry.php
kelunik: Done.

kelunik:
Can we merge this?

kelunik:
@bwoebi @trowski @AndrewCarterUK Can you merge this? I don't want to merge my own PR.

PR #36, opened 2016-05-15
Shorten namespace to Interop\Async and correct PHP version requirement

kelunik:
PR for #34.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,6 @@
 {
     "name": "async-interop/event-loop",
-    "description": "An event loop interface for interoperability.",
+    "description": "An event loop interface",
     "keywords": ["event", "loop", "async", "interop"],
     "license": "MIT",
     "require": {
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -1,14 +1,14 @@
 {
     "name": "async-interop/event-loop",
-    "description": "An event loop interface",
+    "description": "An event loop interface for interoperability.",
```
bwoebi: Interoperability is a _goal_. The project itself is just the interface.

kelunik: That's why there's a _for_. The goal should be in the description.

bwoebi: I still feel a bit uneasy with that. May you please split that into another PR so that others may agree or disagree on it separately?

kelunik: Well, even the package vendor already has `interop` in it.

kelunik: Moved to #38.

PR #37, opened 2016-05-15
Add driver interfaces for lower level access

kelunik:
PR for #16.


assertchris:
I don't understand the need for these. In my mind, `LoopDriver` should be the lowest interface defined. Could you elaborate on the need for them?

kelunik:
They're required for primitives like filesystem access: https://github.com/amphp/file/blob/master/lib/UvDriver.php

assertchris:
Mentioned here (I think): https://github.com/async-interop/event-loop/issues/16

kelunik:
Right, the description already includes a link to #16.

trowski:
-1 on having interfaces for specific implementations. Interfaces should be tied to the features that could be supported. See https://github.com/async-interop/event-loop/issues/16#issuecomment-219310539.

AndrewCarterUK:
I'm -1 here

coderstephen:
Another -1 here... this seems like a step in the wrong direction to me.

PR #46, opened 2016-05-18
Maximum Line Length

AndrewCarterUK:
This is the start of a pull request that could get much longer.

I noticed that the phpdoc descriptions were of a different format (some wrapping at 80 characters some at more).

I've adjusted it so that they all wrap at 80 characters, but I haven't wrapped any @param annotations.

Interested to hear any thoughts.


kelunik:
Should be 120 and we should do something like that at the very end.

AndrewCarterUK:
Good call, I'll close this for now.

PR #59, opened 2016-05-24
Fix running flag, factory reset, etc.

kelunik:
- Fixes resetting the factory by passing `null`.
- Fixes the `$running` flag, which didn't work with nesting so far, thanks @AndrewCarterUK.
- Forbids calling `Loop::setFactory` while in a running loop, as I don't think that's a good idea, because we have that default driver thing to support `wait`.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -17,9 +17,9 @@ final class Loop
     private static $driver = null;
 
     /**
-     * @var bool
+     * @var int
      */
-    private static $running = false;
+    private static $level = 0;
 
     /**
      * Set the factory to be used to create a driver if none is passed to
@@ -30,7 +30,7 @@ final class Loop
     {
         self::$factory = $factory;
 
-        if (!self::$running) {
+        if (self::$level === 0) {
             self::$driver = self::createDriver();
             self::$registry = [];
         }
@@ -53,7 +53,7 @@ final class Loop
 
         self::$driver = $driver;
         self::$registry = [];
-        self::$running = true;
+        self::$level++;
 
         try {
             $callback();
@@ -62,7 +62,7 @@ final class Loop
         } finally {
             self::$driver = $previousDriver;
             self::$registry = $previousRegistry;
-            self::$running = false;
+            self::$level--;
         }
     }
 
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -22,15 +22,26 @@ final class Loop
     private static $level = 0;
 
     /**
-     * Set the factory to be used to create a driver if none is passed to
-     * self::execute. A default driver will be created if none is running
-     * to support synchronous waits in traditional applications.
+     * Set the factory to be used to create a default drivers.
+     *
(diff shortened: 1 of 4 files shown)


bwoebi:
LGTM.

kelunik:
Any other opinions? If we have merged it, I can finally move things to the right namespace.

src/Loop.php
```diff
@@ -17,20 +17,31 @@
...
+        }
+
         self::$factory = $factory;
 
-        if (!self::$running) {
+        if ($factory === null) {
+            self::$driver = null;
+            self::$registry = null;
+        } else {
             self::$driver = self::createDriver();
```
joshdifabio, line 45: Maybe `createDriver()` should be called lazily in `get()` instead of being called in `setFactory()`, in case it's expensive somehow? It was discussed that `setFactory()` could be called by loop implementation packages via composer.json's `autoload.files`, which could mean it sometimes gets called when the driver isn't actually going to be used.

joshdifabio, line 45: `get()` could then simply be changed to:

``` php
return self::$driver ?: self::createDriver();
```

kelunik, line 45: Issue with this is that you can never reset the default driver, maybe we should have a `Loop::reset`?

joshdifabio, line 45: Good point. Could just move `self::$driver = null;` above the if-statement so it's always executed when `setFactory()` is called?

kelunik, line 45: We could also remove it completely and replace it with an explicit `setDefaultDriver`.

bwoebi, line 45: Nah, please don't… @joshdifabio is right with his suggestion, to just set it to null. (When it's null, on the next get() the new instance will be created too).

kelunik, line 45: Having the `$driver` reset with `setFactory` is really weird. I think a separate function is better. It also prevents applications using `wait` from having a default driver at all, that will be different when using `Loop::execute`.

bwoebi, line 45: So, to change the default driver, you'll have to setFactory() _and_ reset() instead of just setFactory()?

kelunik, line 45: No, `setFactory` and `setDefaultDriver` or just `setDefaultDriver` with one argument instead of zero.

kelunik, line 45: But anyway, this is not changed by this PR. It can be changed by another PR later.

bwoebi, line 45: @kelunik but please include @joshdifabio suggestion in _this_ PR.

kelunik, line 45: @bwoebi Why? This can be a separate PR. I don't see a reason why it should be in this PR.

bwoebi, line 45: I said _please_. You don't have to, and I can do it in a separate PR too. The only thing it does is lazy instantiation. Not doing any harm here, it's a mere technical improvement.

kelunik, line 45: Which is a separate thing and separate things should be in separate PRs.

Issue #58, opened 2016-05-23
Operations on Invalid Watchers Must Throw


bwoebi:
From #53:

> May we please add that cancel()'ling multiple times or any other operation on already invalid watchers is disallowed (as watcher ids may be reused)?
> Calling unreference(), reference(), enable() and disable() multiple times must not error.

Is everyone okay with that (I'll open a PR tomorrow to add that in docblocks)? Or are there issues here?

kelunik:
So you want to throw an exception then? We should then also throw exceptions if `unreference` etc. are called with an non-existent watcher.

bwoebi:
For non-existent watchers, sure. Passing these is a mistake which may in edge cases even alter the behavior of the program when the watcher id gets reused.

kelunik:
If you think it's a duplicate of #61, then I'll give this one a better title.

bwoebi:
PR is at #63.

joshdifabio:
Is there any reason `cancel()` can't be idempotent? (I.e. perform a null-op when watcherId doesn't exist instead of throwing.) I can't think of any reason I'd want it to throw in such cases.

bwoebi:
> as watcher ids may be reused

This is the main point. If a watcher id gets reused, you may end up with actually cancelling another watcher than you intended. Thus doubly cancelling or operations on invalid watchers must be a programmer error. (which typically happens spuriously, i.e. if you use spl_object_hash() internally as id; thus very bad.)

joshdifabio:
Is this definitely preferable to requiring that watcher ID's be unique for the duration of the process?

To allow unique watcher ID's across all drivers, the drivers could return watcher ID's which are unique per-driver instance and `Loop` could prepend those ID's with a driver ID (something to uniquely identify a driver instance). This would ensure that the watcher ID's returned by `Loop` were unique for the duration of the process. It could also mean we don't have really long strings (from `spl_object_hash`) being stored all over the place, although I don't know how important that actually is.

Edit: I'm assuming that _drivers could return watcher ID's which are unique per-driver instance_ would be fairly straightforward if a counter was used.

bwoebi:
Actually, I think I'm possibly wrong here...

just use a `static $watcher = "a";` which you can increment on every assign … is unique and the strings are short enough …
After a _hundred trillion_ of watchers you still only have 10 chars … and 10 chars will probably cover the life-time of a server… (i.e. 11 chars are enough for 10 million watchers per second for 11 years.)

[...]

PR #63, opened 2016-05-26
Throw on invalid watcher identifier

trowski:
Addresses #58. Adds `InvalidWatcherExeption` that should be thrown if `enable`, `disable`, `cancel`, `reference`, or `unreference` is attempted with an invalid watcher identifier.

### DIFF
--- a/src/Loop/InvalidWatcherException.php
+++ b/src/Loop/InvalidWatcherException.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Interop\Async\Loop;
+
+/**
+ * Must be thrown if an operation is attempted with an invalid watcher identifier.
+ */
+class InvalidWatcherException extends \RuntimeException
+{
+
+}
--- a/src/Loop/InvalidWatcherException.php
+++ b/src/Loop/InvalidWatcherException.php
@@ -5,7 +5,7 @@ namespace Interop\Async\Loop;
 /**
  * Must be thrown if an operation is attempted with an invalid watcher identifier.
  */
-class InvalidWatcherException extends \RuntimeException
+class InvalidWatcherException extends \LogicException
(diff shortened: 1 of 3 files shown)


bwoebi:
I'd explicitly specify _invalid or cancelled_. This eliminates every ambiguity.
Cancelling may or may not invalidate the watcher internally immediately. The guarantee is that it will be invalidated after the next tick and will not be called after cancellation anymore.

src/Loop/InvalidWatcherException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Interop\Async\Loop;
+
+/**
+ * Must be thrown if an operation is attempted with an invalid watcher identifier.
+ */
+class InvalidWatcherException extends \RuntimeException
```
bwoebi: This probably should be a LogicException as this always is a programmer error.

trowski: Oops, I meant for it to be a LogicException, thanks.

bwoebi:
Do we maybe want to explicitly add (on Driver interface) that `disable()` _never_ invalidates a watcher?

trowski:
I agree, I'll add that as well.

kelunik:
> Cancelling may or may not invalidate the watcher internally immediately. 

I'd say it should do that immediately.

bwoebi:
@kelunik yes, _should_. But that's an implementation detail we have no control of.

kelunik:
It's something we can define. Then we have the control.

bwoebi:
It may eventually depend on the loop implementation. I.e. uv internally only releases it on the next tick. (to avoid freed memory access; see also uv_close() on libuv docs)

PR #65, opened 2016-05-26
Lazy factory instantiation and save method call

bwoebi:
The loop code is quite hot, thus it should not be too bad to save a method call here. `self::get()` is still called to eventually create&assign the default driver, if it is missing.

Also instantiate the factories lazily. This additionally avoids having an unused default loop object in many cases.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -40,19 +40,14 @@ final class Loop
         }
 
         self::$factory = $factory;
-
-        if ($factory === null) {
-            self::$driver = null;
-        } else {
-            self::$driver = self::createDriver();
-        }
+        self::$driver = null; // reset it here, it will be actually instantiated inside execute() or get()
     }
 
     /**
      * Execute a callback within the scope of an event loop driver.
      *
      * @param callable $callback The callback to execute
-     * @param Driver $driver The event loop driver
+     * @param Driver $driver The event loop driver. If null a new one is created from the set factory.
      *
      * @return void
      */
@@ -60,9 +55,7 @@ final class Loop
     {
         $previousDriver = self::$driver;
 
-        $driver = $driver ?: self::createDriver();
-
-        self::$driver = $driver;
+        self::$driver = $driver ?: self::createDriver();
         self::$level++;
 
         try {
@@ -103,11 +96,10 @@ final class Loop
      */
     public static function get()
     {
-        if (null === self::$driver) {
-            throw new \RuntimeException('Missing driver; Neither in Loop::execute nor factory set.');
+        if (self::$driver) {
+            return self::$driver;
         }
-
-        return self::$driver;
+        return self::$driver = self::createDriver();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Loop.php
```diff
@@ -42,12 +42,11 @@ public static function setFactory(DriverFactory $factory = null)
...
 
         self::$factory = $factory;
+        self::$driver = null; // reset it here, it will be actually instantiated inside execute() or get()
 
         if ($factory === null) {
-            self::$driver = null;
             self::$registry = null;
         } else {
-            self::$driver = self::createDriver();
             self::$registry = [];
```
kelunik: Should we still have the registry here then?

src/Loop.php
```diff
@@ -124,7 +120,8 @@ public static function get()
      */
     public static function stop()
     {
-        self::get()->stop();
+        $driver = self::$driver ?: self::get();
```
kelunik, line 112: You can as well just directly call `self::createDriver()` instead of `self::get()`.

src/Loop.php
bwoebi: Yes, else we get problems when we assign to registry _before_ the first watcher creation.

src/Loop.php
bwoebi, line 112: Haha, I did the same mistake before. `self::createDriver()` just creates it and returns it, but doesn't assign it.

kelunik, line 112: Just use `self::$driver = self::$driver ?: self::createDriver();`

bwoebi, line 112: Now you're microdeoptimizing :-P Assigns to/reads from static props are more expensive!

src/Loop.php
kelunik: We have an issue there anyway, because the registry isn't reused for wait currently. It's always a fresh one.

https://github.com/async-interop/event-loop/blob/2c22efbb8a4b01fdec3928842397fb1bd0a842ea/src/Loop.php#L62-L80

bwoebi: Okay, that's a bug then.

kelunik: We'd need a getter for the complete registry state. I really don't like that. Stupid `wait` ...

bwoebi: This issue isn't dependent on `wait()` specifically, but rather generically whenever you reuse an event loop.

E.g.:

``` php
$loop = new DriverInstanceFromSomeWhere;
Loop::execute(function() {
    Loop::storeState("foo", 1);
}, $loop);
Loop::execute(function() {
    var_dump(Loop::fetchState("foo")); // ugh???
}, $loop);
```

kelunik: Yes, sure. But the only reason to do so is `wait`. Anything other shouldn't reuse a loop. If we didn't have wait, we could explicitly forbid that. Maybe leave it as trait and make it a must for `LoopDriver`?

bwoebi: There may be reasons to reuse the loop. I.e. an exception was thrown and you handled it outside. The loop shall be re-runnable then.

Making it a trait for usage in Drivers is an option. Not sure if it's the best one, but it isn't bad.

kelunik: Better than a getter for the complete registry and passing that one in as well.

PR #67, opened 2016-05-26
Put Loop::execute() $callback into Driver::defer()

bwoebi:
This also ensures that potential throwing in the callback will be handled inside a potential error handler instead of immediately falling through.

P.s.: Test changes are necessary as it now is in defer()... Thus need to execute the defers...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Interop\\Async\\Loop\\": "test"
+            "Interop\\Async\\Loop\\Test\\": "test"
         }
     }
 }
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -72,8 +72,7 @@ final class Loop
         self::$level++;
 
         try {
-            $callback();
-
+            self::$driver->defer($callback);
             self::$driver->run();
         } finally {
             self::$driver = $previousDriver;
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Interop\\Async\\Loop\\": "test"
+            "Interop\\Async\\Loop\\Test\\": "test"
```
kelunik, line 19: Is there an actual reason to have a separate test namespace?

bwoebi, line 19: Yeah, because there's now also the Dummy-Driver, and not to be confused with the event-loop-tests repo

kelunik, line 19: > and not to be confused with the event-loop-tests repo

How's that related? That one should have its own namespace.

bwoebi, line 19: event-loop-tests currently provides one single class Interop\Async\Loop\Test.

But still, alone because of the DummyDriver it ought to not be in the main Loop namespace.

kelunik, line 19: > But still, alone because of the DummyDriver it ought to not be in the main Loop namespace.

And why? There won't be a collision.

bwoebi, line 19: No, there's not. Is there any argument _against_ the namespace though?

kelunik, line 19: Needs more use statements.

bwoebi, line 19: That's not true. It only needs to access Loop class, which anyway needs to be imported.

This PR does not add any use statement. Or am I missing anything?

PR #69, opened 2016-05-28
(Re-)Introduce InvalidWatcherException

bwoebi:
As discussed in #58, this PR proposes to only throw on enable() if the watcher has been cancel()'led.

Note that the other functions still have a `@throws` annotation; a loop driver must throw if (un-)reference(), cancel() or disable() are attempted with an invalid watcher identifier (i.e. any identifier not yet emitted by the driver, thus never has been valid in the whole lifetime of the driver).


kelunik:
This is a nogo. It can't have a MAY throw. Either MUST or MUST NOT. Additionally, there should be defined what's unique. Unique to the loop? Unique to the currently existing ones? Unique to the program? Unique in the whole world?

bwoebi:
Currently unique to the loop. Another PR may upgrade it to unique to the program.

Also, why is MAY a problem? It depends on whether the loop has a way to keep track of it (i.e. if they're incremental, it's easy, do a compare).
This one is more like a hint "there is a programmer error, but it actually doesn't impact the program".

kelunik:
And then it impacts a program, because it just works in the one and throws with another loop. The set of thrown exceptions must be very clearly defined.

bwoebi:
I don't agree, but I actually don't see any actual arguments against MUST as loops just should use an incremental id which is easy to verify. Changed it.

bwoebi:
@kelunik Note: If we enforce unique to the program, we cannot verify easily anymore whether an identifier has ever been valid _for the driver instance_.

Thus it's probably better to have it unique per driver instance.

kelunik:
The advantage of unique to the program and always throwing would be that objects used within multiple loops can't ever cancel watchers they didn't indent to.

bwoebi:
If you have an implementation how to solve that without O(n) storage of all ever emitted watchers of all the drivers, you're welcome.

kelunik:
Why do you need a storage of all ever emitted watchers?

bwoebi:
To verify whether the watcher has been emitted by this specific driver instance. Or how would you verify this? (assuming cancel() etc. all remain idempotent, which they definitely should)

kelunik:
Why should cancel be idempotent? Why should I have to verify if a specific driver instance created the watcher? If it doesn't exist, it just throws.

bwoebi:
@kelunik We've already discussed this here: https://github.com/async-interop/event-loop/issues/58#issuecomment-222154656

kelunik:
How often does that really happen? Usually you store the watcher as a property anyway.

bwoebi:
How often does it happen that you stop the loop twice?
Well, we don't know. We can just lift the restrictions where possible.

And, yes, I've already called cancel() multiple times. On a property. I.e. first the internal onReadable() cancelled it and then a manual close() function call.

kelunik:
> How often does it happen that you stop the loop twice?

You don't keep references to the loop. You don't know whether you stopped it or not. For watchers, you normally store a property and clear it as soon as you cancel the watcher.

trowski:
@kelunik Allowing multiple calls to `cancel`, `disable`, etc. reduces the need for extra logic and status variables when using watchers. The only operation that should throw on an invalid watcher is `enable`, as this would change intended behavior. Double disabling or cancelling a watcher should just be a no-op.

I'm not sure throwing on an invalid watcher ID (one that has not been issued by the loop) in any method other than `enable` is necessary. If using an incremental watcher ID this is easy to check, but I feel like we're then dictating too much about how watcher IDs must be generated, as well performing yet another check on every watcher op.

kelunik:
@trowski I don't understand why it's OK to have `disable` called multiple times, but not `enable`. Could you explain again?

bwoebi:
enable() changes the state. disable() on something cancelled is a no-op [cancel disables and removes basically, thus it's idempotent]. There is no state change expected.

trowski:
@kelunik Calling `enable` multiple times on a _valid_ watcher should be fine, but calling `enable` on a cancelled watcher is a mistake.

bwoebi:
Note that I'm also fine with not throwing at all for anything but enable(). But enable() on invalid/cancelled() watchers MUST throw in any case, and cancel() etc. MUST definitely not throw for previously valid watchers.

kelunik:
> enable() changes the state. disable() on something cancelled is a no-op [cancel disables and removes basically, thus it's idempotent]. There is no state change expected.

If `enable` throws on invalid watchers, `disable` has to throw as well. I don't see why they should be different.

bwoebi:
Definitely not. The property of `enable()` is to enable the watcher. If there is none, that's bad.
The common property of `disable()` and `cancel()` is to change state to disabled (first). The watchers do not exist i.e. they were disabled sometime before, thus there's no state change necessary, thus no reason to throw. [Idempotent.]

kelunik:
If you enable multiple times, that's the same thing. It's mostly a programming error if you disable a non-existent watcher.

bwoebi:
Enabling multiple times is no problem. It doesn't invoke any state change.

If there were some sort of invalid always-enabled watcher, enabling that one multiple times would obviously work and disable() wouldn't. But that's not the case.

kelunik:
If one throws, the other should throw as well. Disabling is not canceling. I'm fine if canceling works different, but disable must work just as enable. It has to be consistent IMO.

bwoebi:
Right, cancel() is close in semantics to disable(). Disabling and enabling are very different things though. The former two need to be consistent, not necessarily the latter two.

kelunik:
I disagree there then.

bwoebi:
Thus, as we fundamentally disagree, shall we just vote on the matter how others see it?

kelunik:
Sure, cancel includes disable, to that's why you say the semantics are similar, but cancel is just a superset  here. Enable and disable are exact opposites and thus should behave the same IMO (regarding invalid watchers and multi calls).

bwoebi:
cancel() is a _strict_ superset and it should behave exactly like disable, with the additional functionality that it cannot be re-enabled after that.

And that's why subsequent enable()'s must fail. It's the explicit functionality of cancel() to prevent that.
It does not explicitly prevent subsequent disables, so that's not an issue.

kelunik:
> It's the explicit functionality of cancel() to prevent that.

No, the explicit functionality of cancel is to delete the watcher, so it can neither be enabled, disabled, referenced nor unreferenced. Disabling is just a detail.

AndrewCarterUK:
My thoughts:

| Call | State | Result |
| --- | --- | --- |
| enable | enabled | success |
| enable | disabled | success |
| enable | cancelled/invalid | exception |
| disable | enabled | success |
| disable | disabled | success |
| disable | cancelled/invalid | exception |
| cancel | enabled | success |
| cancel | disabled | success |
| cancel | cancelled/invalid | exception |

bwoebi:
@AndrewCarterUK would you please provide your reasoning (to see whether it's actually reasonable and we just disagree or whether you have a flaw in your logic)?

AndrewCarterUK:
Apologies, reasoning is that the invalid watcher exception should be thrown for invalid watchers. Cancelled is the process of making a watcher invalid (cancelled watcher == invalid watcher), so if you then proceed to use that watcher you should raise the invalid watcher exception.

I don't see enabling an enabled watcher as an exceptional event, though we could use the return value to indicate if there was a state change if people wished (I wouldn't bother).

src/Loop/Driver.php
```diff
@@ -105,6 +105,8 @@ public function onSignal($signo, callable $callback, $data = null);
      * @param string $watcherId The watcher identifier.
      *
      * @return void
+     *
+     * @throws InvalidWatcherException Thrown if the watcher identifier is invalid or cancelled.
```
AndrewCarterUK, line 109: Invalid = cancelled?

bwoebi:
No it isn't exceptional. It's just (IMO) exceptional to enable a watcher which was marked as non-enablable through cancel().

(That cancel() actually may unset() the watcher in the loop is an implementation detail to me.)

src/Loop/Driver.php
bwoebi, line 109: > Invalid watcher identifier: any identifier not yet emitted by the driver

kelunik, line 109: I still don't see a reason to differentiate that.

kelunik:
Maybe we should just call `cancel` → `delete` then, so it's clear what it does? Does not just disable and prevent re-enable?

src/Loop/Driver.php
bwoebi, line 109: I do not care too much about this; this just was part of the initial proposal and I was explaining to Andrew as he asked.

AndrewCarterUK:
I'm not sure I'm following, are you saying that we shouldn't throw an `InvalidWatcherException` if someone passes an invalid watcher to `disable()` or `cancel()`?

There is possibly some debate around `cancel()`. I could see why we might not want to throw an exception there - but if you call the `disable()` method that is explicitly because you might want to re-enable it (otherwise you'd have called `cancel()`) so that is certainly an exceptional event.

src/Loop/Driver.php
kelunik, line 109: Which initial proposal? Of this PR?

bwoebi:
@kelunik no, please not. That's not what I want it to do. No.

bwoebi:
@AndrewCarterUK right, that's the discussion.

And the question is whether that's exceptional. I _can_ get on board with disable (you typically only disable on\* watchers in their direct callback...)…

AndrewCarterUK:
Well if you call the `disable` method with a watcher, you expect the watcher to be in a disabled state after you have done so. If the watcher isn't disabled (it's cancelled), there is definitely an exception. I honestly don't think there's any discussion on that one.

With regards to calling `cancel`, the arguments for throwing the exception:
- Creates an inconsistency with `enable()` and `disable()` as they throw an invalid watcher exception when given invalid watcher

Arguments against:
- Creates an inconsistency with `enable()` and `disable()` as they both allow enabling enabled watchers and disabling disabled watchers.

Anyone add any more, I could be persuaded either way at the moment?

src/Loop/Driver.php
AndrewCarterUK, line 109: I was just commenting that the PHPDoc is inconsistent with the other methods (which assume that invalid and cancelled are the same).

bwoebi:
You may argue though that there is only enabled and disabled. That cancel() sets it to disabled() and sets a flag that it cannot be re-enabled.
That's my argumentation.

I still think I am correct, but I am willing to find a compromise here.

But I really think that multiple cancel() must work, because, as you say:

> - Creates an inconsistency with enable() and disable() as they both allow enabling enabled watchers and disabling disabled watchers.

You even may argue it isn't inconsistent with `enable()` and `disable()` as these _always_ throw an exception if they encounter a state they cannot change (i.e. cancelled, if you consider that a state). (But cancelling a cancelled watcher is not a state change, thus allowed.)

bwoebi:
Closing this, I'll open a new PR, throwing on invalid watchers for enable/disable in general, and allowing multiple cancel.

kelunik:
> You may argue though that there is only enabled and disabled. That cancel() sets it to disabled() and sets a flag that it cannot be re-enabled.

So you argue that watchers are essentially never deleted, just not enableable anymore?

bwoebi:
From the outside perspective, yes.

I'm saying this is a possible view on it. Not that it's necessarily the correct view.

kelunik:
In that case we would have multiple methods creating watchers but none to ever delete it. I don't think that view makes that much sense.

bwoebi:
You don't need to delete it, thus we don't need to provide a method to actually delete it either.

But anyway, going to open an other PR with multiple cancel() allowed, but no other operations on invalidated watchers.

kelunik:
> You don't need to delete it, thus we don't need to provide a method to actually delete it either.

You need to delete it if you want to completely free it. So cancel is essentially delete. Maybe we should even call it delete then.

bwoebi:
Actually freeing (aka releasing the memory) is the responsibility of the implementation.

AndrewCarterUK:
Thinking about this I think it makes sense to rename `cancel()` to `delete()` to make it more expressive and obvious.

If I'm building a long running server application, I **need** to have some option for freeing all of the resources associated with a watcher. Without this, I can't guarantee that my application doesn't have a memory leak - so it is a requirement.

bwoebi:
Well, actually, … I even added a test against memory leaking: https://github.com/async-interop/event-loop-test/blob/de3c259557a0581ba2e0edd8225bf6a25980822a/src/Test.php#L373

But this still is the drivers responsibility. I'm open to adding this as an explicit requirement to the _Driver::cancel()_. But it really isn't the business of the Loop. The Loop is just an easy facade for instructions.

AndrewCarterUK:
I don't think it's possible to build a long running application using this standard if there is no method that can be used to free the resources associated with watchers.

I'm open to any solution which solves this problem, but I think it is the business of the standard - as without some guarantee of this form, the standard is useless for anything other than very short lived applications.

kelunik:
> The Loop is just an easy facade for instructions

Yes, all driver instructions. Or are you talking about the loop driver vs. underlying loop?

PR #72, opened 2016-06-02
(Re-)Introduce InvalidWatcherException

### LINKED ISSUE #58: Operations on Invalid Watchers Must Throw

bwoebi:
Fixes #58.

Followup to #69 with the following rules:

Any operation must happen on a valid watcher, except cancel(), which does not check watcher validity (and just silently returns).

### DIFF
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -33,7 +33,7 @@ interface Driver
      * @param callable(string $watcherId, mixed $data) $callback The callback to defer.
      * @param mixed $data Arbitrary data given to the callback function as the $data parameter.
      *
-     * @return string An identifier that can be used to cancel, enable or disable the watcher.
+     * @return string An unique identifier that can be used to cancel, enable or disable the watcher.
      */
     public function defer(callable $callback, $data = null);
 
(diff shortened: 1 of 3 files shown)


src/Loop/Driver.php
```diff
@@ -114,11 +116,13 @@ public function enable($watcherId);
...
      *
      * @return void
+     *
+     * @throws InvalidWatcherException Thrown if the watcher identifier is invalid.
      */
     public function disable($watcherId);
 
     /**
-     * Cancel a watcher.
+     * Cancel a watcher. This marks the watcher as invalidated. Calling this function MUST never fail, even when passed an invalid watcher.
```
kelunik: `invalidated` → `invalid` to be consistent.

src/Loop/InvalidWatcherException.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+namespace Interop\Async\Loop;
+
+/**
+ * MUST NOT be thrown by cancel().
```
kelunik: That shouldn't be mentioned here.

bwoebi:
May everyone please add a :+1: or add a comment with his problems so that we can eventually merge it if there are no complaints.

AndrewCarterUK:
Might be worth being explicit that cancel should not throw an exception when given an invalid watcher if this is the route we are going down.

bwoebi:
It's present on the Driver::cancel docs:

> - Cancel a watcher. This marks the watcher as invalid. Calling this function MUST never fail, even when passed an invalid watcher.

It already has no @throws annotation, that should be good enough for Loop::cancel() docs.

PR #74, opened 2016-06-04
Specify exact guarantees when a callback has to be called

bwoebi:
This prevents unexpected failures due to different execution orders in different implementations.

### DIFF
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -28,7 +28,7 @@ interface Driver
     /**
      * Defer the execution of a callback.
      *
-     * The deferred callable MUST be executed in the next tick of the event loop.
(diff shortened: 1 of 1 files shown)


src/Loop/Driver.php
```diff
@@ -40,7 +40,7 @@ public function defer(callable $callback, $data = null);
     /**
      * Delay the execution of a callback.
      *
-     * The delay is a minimum and approximate, accuracy is not guaranteed.
+     * The delay is a minimum and approximate, accuracy is not guaranteed. Order of calls MUST be determined by which timers expire first.
```
kelunik: Maybe add something if time is equal. Then order decides.

bwoebi: That was implied here (one delay always expires first as it is declared a few microseconds earlier...)

How would you word it?

kelunik: As we don't have microsecond resolution, just millisecond, the first scheduled one should be executed first.

PR #87, opened 2016-08-04
Merge Registry into Driver

kelunik:
See #85.


src/Loop/Driver.php
```diff
@@ -13,7 +18,7 @@
      *
      * @return void
      */
-    public function run();
+    public abstract function run();
```
joelwurtz: it should be `abstract public` for PSR2 compliance (same for others of course :) )

kelunik: I hate PSR-2.

geggleto: agreed.

kelunik: Fixed in the latest commit.

src/Loop/Driver.php
```diff
@@ -189,19 +194,29 @@ public function unreference($watcherId);
      *
      * @return void
      */
-    public function storeState($key, $value);
+    public final function storeState($key, $value)
```
joelwurtz: I think you will hate me, but this is also not PSR2 compliant (should be `final public`), i'm sorry, but if one day if see you in person i will offer you a beer to compensate :)

(Same for fetchState)

kelunik: Why should I hate you? It's only PSR-2. Fixed, thanks. :-)

Actually, I remembered I should do this, but forgot to do so after I opened the editor.

PR #91, opened 2016-08-25
Allow disabling of invalid watchers

### LINKED ISSUE #90: Throwing from disable() with invalid watcher
I've been running into timing issues causing `Loop::disable()` to throw due to a watcher being invalidated during the invocation of the watcher callback. Essentially, if the watcher callback calls `Loop::disable()` at the end of the callback after side-effects from the watcher callback have invoked `Loop::cancel()`, an exception is being thrown. While it is possible to use an additional flag to code around this problem, it seems unnecessary.

Some example code: https://github.com/amphp/socket/blob/amp_v2/lib/Server.php#L37-L51

Putting the [call to `Loop::cancel()` in a defer watcher](https://github.com/amphp/socket/blob/amp_v2/lib/Server.php#L76-L78) _seems_ to solve the issue here. I tried the same in the `Socket` class within the same lib [here](https://github.com/amphp/socket/blob/amp_v2/lib/Socket.php#L200-L203). However, this does not appear to have solved the problem, as tests still occasionally result in [throws due to invalid watchers](https://travis-ci.org/amphp/concurrent/jobs/153426676#L418).

I propose that `InvalidWatcherException` should not be thrown from `disable()` with cancelled watchers, as disabling a cancelled watcher does not change the expectation that further events will not be received on the watcher. `enable()` should continue throwing when a cancelled watcher is given.

bwoebi:
Fixes #90.

Following no dissent in the related issue, this should be merged in a few days if nobody else objects.

### DIFF
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -134,19 +134,18 @@ abstract class Driver
     abstract public function enable($watcherId);
 
     /**
-     * Disable a watcher. Disabling a watcher MUST NOT invalidate the watcher.
+     * Disable a watcher. Disabling a watcher MUST NOT invalidate the watcher. Calling this function MUST NOT fail,
+     * even if passed an invalid watcher.
      *
      * @param string $watcherId The watcher identifier.
      *
      * @return void
-     *
-     * @throws InvalidWatcherException If the watcher identifier is invalid.
      */
     abstract public function disable($watcherId);
 
     /**
      * Cancel a watcher. This will detatch the event loop from all resources that are associated to the watcher. After
-     * this operation the watcher is permanently invalid. Calling this function MUST never fail, even when passed an
+     * this operation the watcher is permanently invalid. Calling this function MUST NOT fail, even if passed an
      * invalid watcher.
      *
      * @param string $watcherId The watcher identifier.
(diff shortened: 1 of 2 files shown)


src/Loop/Driver.php
```diff
@@ -134,13 +134,12 @@
     abstract public function enable($watcherId);
 
     /**
-     * Disable a watcher. Disabling a watcher MUST NOT invalidate the watcher.
+     * Disable a watcher. Disabling a watcher MUST NOT invalidate the watcher. Calling this function MUST never fail,
```
kelunik: `MUST never` → `MUST NOT`

src/Loop/Driver.php
```diff
@@ -134,13 +134,12 @@
...
+     * even when passed an invalid watcher.
```
kelunik: I guess this should be `if`, not `when`.

src/Loop/Driver.php
bwoebi: I've copied from cancel() docs.

kelunik: Should probably fixed there as well. Otherwise you might miss the `never` since only `MUST` is bold.

PR #100, opened 2016-09-23
Clarify that (half-)closed streams also trigger the respective read and write watchers

### LINKED ISSUE #99: Watcher behavior regarding resources closed unexpectedly
What is the expected behavior of read / write watchers when the watched resource is closed unexpectedly (broken pipe / remote peer closed a socket...)?

The doc comments in the Loop class read like this:

> Execute a callback when a stream resource becomes readable / writable.

I implemented loop drivers based on stream_select(), ev and event. All of them report broken resources as readable / writable. Same thing with uv, it triggers polls with an empty event mask indicating a broken resource. I think the docs should be updated to include this case and define it as expected behavior.

bwoebi:
Fixes #99.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -172,7 +172,7 @@ final class Loop
     }
 
     /**
-     * Execute a callback when a stream resource becomes readable.
+     * Execute a callback when a stream resource becomes readable or is closed for reading.
      *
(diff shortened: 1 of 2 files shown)


kelunik (approved):
LGTM.

PR #101, opened 2016-09-24
Fix code markdown for Loop::info

kelunik:


bwoebi:
What's the _fix_ here? I.e. why does this need to be indented by four chars??

kelunik:
Because PHPDoc is supposed to be markdown and it's supposed to be formatted as code in docs.

kelunik:
If it's not formatted as code, any markdown parser will show it like that:

image

kelunik:
See #103.

PR #109, opened 2016-10-24
Add warning about locally closed resources being undefined behavior

### LINKED ISSUE #106: Streams closed before watcher is enabled
Since read / write watchers are enabled after the current tick of the event loop there is a possibility that a stream resource is closed by another watcher within the same tick. This causes trouble with many loop backends (ev, event, uv) because they fail at creating new event watchers when the stream resource has been closed.

This is an edge case but i think that the specification should be updated to define expected behavior in this case. It would be sufficient to have loop implementations check for valid resources as new stream watchers are enabled for the first time and call the watcher callback once if a closed / invalid resource is detected.

kelunik:
Also makes Loop docs consistent with Driver. Resolves #106.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -178,11 +178,18 @@ final class Loop
     /**
      * Execute a callback when a stream resource becomes readable or is closed for reading.
      *
+     * Warning: Closing resources locally, e.g. with `fclose`, might not invoke the callback. Be sure to `cancel` the
+     * watcher when closing the resource locally. Drivers might choose to notify the user in a debug mode if there are
+     * watchers on invalid resources, but are not required to, due to the high performance impact. Watchers on closed
+     * resources are therefore undefined behavior.
+     *
+     * Multiple watchers on the same stream may be executed in any order.
+     *
      * @param resource $stream The stream to monitor.
      * @param callable(string $watcherId, resource $stream, mixed $data) $callback The callback to execute.
      * @param mixed $data Arbitrary data given to the callback function as the `$data` parameter.
      *
-     * @return string An identifier that can be used to cancel, enable or disable the watcher.
+     * @return string An unique identifier that can be used to cancel, enable or disable the watcher.
      */
     public static function onReadable($stream, callable $callback, $data = null)
     {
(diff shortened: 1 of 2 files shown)


src/Loop.php
```diff
@@ -178,11 +178,18 @@ public static function repeat($interval, callable $callback, $data = null)
     /**
      * Execute a callback when a stream resource becomes readable or is closed for reading.
      *
+     * Warning: Closing resources locally, e.g. with `fclose`, might not invoke the callback. Be sure to `cancel` the
+     * watcher when closing the resource locally. Drivers might choose to notify the user in a debug mode if there are
```
bwoebi: s/might choose to notify the user in a debug mode/may choose to notify the user

There's no need for talking about a debug mode - that's an implementation detail unrelated to the specification here.

bwoebi (changes requested):

src/Loop.php
kelunik: There's no _need_, but just a hint.

bwoebi: The emphasis is on removing "in a debug mode". Just keep it out, it has no business here.
Maybe put it in a meta document outlining how you intend it to be used, but it's not appropriate here.

bwoebi (approved):

WyriHaximus (approved):
:+1:

src/Loop/Driver.php
```diff
@@ -81,9 +81,9 @@
      * Execute a callback when a stream resource becomes readable or is closed for reading.
      *
      * Warning: Closing resources locally, e.g. with `fclose`, might not invoke the callback. Be sure to `cancel` the
-     * watcher when closing the resource locally. Drivers might choose to notify the user in a debug mode if there are
-     * watchers on invalid resources, but are not required to, due to the high performance impact. Watchers on closed
-     * resources are therefore undefined behavior.
+     * watcher when closing the resource locally. Drivers may choose to notify the user if there are watchers on invalid
```
kelunik: Maybe we should write `MAY`

kelunik:
Forgot to copy `defer`, `delay`, `repeat` from `Driver` to `Loop`. Also replaced `may` with `MAY` and rebased onto `master`.

bwoebi:
Feel free - would be consistent with the other text.

Bob Weinand (iPhone)

kelunik:
That's a bit late.

PR #135, opened 2017-01-08
watchers → enabled_watchers in getInfo

kelunik:
This ensures it's clear which info is returned. I could have added another sentence to the docblock instead, but I think this makes it more obvious when consuming the code, too.

In general, I think that we should be careful with allowing extension here, as it limits us in adding more specified keys, such as `"disabled_watchers"`.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -402,13 +402,13 @@ final class Loop
      * The returned array MUST contain the following data describing the driver's currently registered watchers:
      *
      *     [
-     *         "defer"         => ["enabled" => int, "disabled" => int],
-     *         "delay"         => ["enabled" => int, "disabled" => int],
-     *         "repeat"        => ["enabled" => int, "disabled" => int],
-     *         "on_readable"   => ["enabled" => int, "disabled" => int],
-     *         "on_writable"   => ["enabled" => int, "disabled" => int],
-     *         "on_signal"     => ["enabled" => int, "disabled" => int],
-     *         "watchers"      => ["referenced" => int, "unreferenced" => int],
+     *         "defer"            => ["enabled" => int, "disabled" => int],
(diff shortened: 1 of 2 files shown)


kelunik:
This is a PR for #131.

bwoebi (approved):
I'll merge tomorrow if no-one else objects.

kelunik:
@bwoebi: What do you think about limiting extension to allow for future extension in the specification?

bwoebi:
Once we merge the separate virtual package for API, this isn't an issue.
Yes, it will require a major version for the Driver API, but not for the loop implementations themselves. (unless the implementations define a key with the same name but different value; in that case semver following implementations have to release a major version - but that's then specific for the applications implementing that particular loop.)

I do not see that future extension is exceedingly limited here.

kelunik:
But those implementations have to release a new major version. IMO they should add a separate method if they want to add more data,  that will have a way lower chance of conflicting. Additionally, this would make the return of `getInfo` completely predictable.

bwoebi:
Well, it's the choice of the implementations whether they choose to add more entries and incur the risk of having to sometime release a new major in the unlikely case of same key, different value.

The same way round you may argue that you could - without this restriction - rename keys in the Driver without requiring the implementations to release a new major (because they're allowed to retain the old key then).

kelunik:
> The same way round you may argue that you could - without this restriction - rename keys in the Driver without requiring the implementations to release a new major (because they're allowed to retain the old key then).

No, you can't. This would break the API. I'm specifically talking about _not_ breaking the consumer API, but still being able to break the SPI.

kelunik:
@trowski @rdlowrey @WyriHaximus How do you think about the extension?

WyriHaximus (approved):
Makes perfect sense to me :+1:

kelunik:
Could anyone merge this?

PR #137, opened 2017-01-08
Enforce unique-per-process watcher identifiers

joshdifabio:
This PR aims to enforce unique-per-process watcher identifiers. It does so in a relatively cheap way, adding a single function call per watcher created, and also provides a more expensive validation method which drivers should only call when they do not recognise a watcher identifier. This validation method is able to distinguish between the following error cases:
- ~Watcher identifier invalid~
- Watcher identifier passed to wrong driver instance
- ~Watcher identifier no longer valid~

Edit: The PR is now more efficient but provides less fine-grained errors.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -295,11 +295,14 @@ final class Loop
      * Cancel a watcher.
      *
      * This will detatch the event loop from all resources that are associated to the watcher. After this operation the
-     * watcher is permanently invalid. Calling this function MUST NOT fail, even if passed an invalid watcher.
+     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by this
+     * driver but is no longer valid.
      *
      * @param string $watcherId The watcher identifier.
      *
      * @return void
+     *
+     * @throws InvalidWatcherException If the watcher identifier is invalid.
      */
     public static function cancel($watcherId)
     {
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -295,14 +295,14 @@ final class Loop
      * Cancel a watcher.
      *
      * This will detatch the event loop from all resources that are associated to the watcher. After this operation the
-     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by this
-     * driver but is no longer valid.
+     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by the
+     * current driver but is no longer valid.
      *
      * @param string $watcherId The watcher identifier.
      *
      * @return void
      *
-     * @throws InvalidWatcherException If the watcher identifier is invalid.
+     * @throws InvalidWatcherException If the watcher identifier does not belong to the current driver.
      */
     public static function cancel($watcherId)
     {
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -17,12 +17,12 @@ abstract class Driver
     /**
      * @var string
      */
-    private $driverId;
+    private $watcherIdPrefix;
 
     /**
      * @var string
      */
-    private $nextWatcherId = 'a';
+    private $nextWatcherId;
 
     /**
      * @var array
@@ -34,7 +34,8 @@ abstract class Driver
      */
     public function __construct()
     {
-        $this->driverId = self::$nextDriverId++ . '-';
+        $this->watcherIdPrefix = self::$nextDriverId++ . '-';
+        $this->nextWatcherId = "{$this->watcherIdPrefix}a";
     }
 
     /**
@@ -322,7 +323,7 @@ abstract class Driver
      */
     final protected function createWatcherId()
     {
-        return $this->driverId . $this->nextWatcherId++;
+        return $this->nextWatcherId++;
     }
 
     /**
@@ -338,9 +339,9 @@ abstract class Driver
      *
      * @throws InvalidWatcherException
      */
-    final protected function validateWatcherId($watcherId, bool $mustThrow)
+    final protected function validateWatcherId($watcherId, $mustThrow)
     {
-        if (0 !== \strpos($watcherId, $this->driverId)) {
+        if (0 !== \strpos($watcherId, $this->watcherIdPrefix)) {
             // the watcher was created by a different driver
             throw new InvalidWatcherException($watcherId, 'A watcher was passed to the wrong driver.');
         }
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -326,7 +326,7 @@ abstract class Driver
     }
 
     /**
-     * Validate a watcher identifier which was passed to the driver by a client.
+     * Validate a watcher identifier which has been passed to the driver by a client.
      *
      * The driver MUST call this method when it is provided with a watcher identifier which it does not recognise. The
      * driver SHOULD NOT call this method every time it receives a watcher identifier, as doing so would be relatively
@@ -339,7 +339,11 @@ abstract class Driver
     final protected function validateWatcherId($watcherId)
     {
         if (0 !== \strpos($watcherId, $this->watcherIdPrefix)) {
-            throw new InvalidWatcherException($watcherId, 'A watcher was passed to the wrong driver.');
+            throw new InvalidWatcherException(
+                $watcherId,
+                "Watcher '{$watcherId}' has been passed to the wrong driver. Watchers passed to this driver must " .
+                "start with '{$this->watcherIdPrefix}'."
+            );
         }
     }
 }
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -180,12 +180,14 @@ abstract class Driver
      * Cancel a watcher.
      *
      * This will detach the event loop from all resources that are associated to the watcher. After this operation the
-     * watcher is permanently invalid. Calling this function MUST NOT fail even if the passed watcher was issued by this
+     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by this
      * driver but is no longer valid.
      *
      * @param string $watcherId The watcher identifier.
      *
      * @return void
+     *
+     * @throws InvalidWatcherException If the watcher identifier is invalid.
      */
     abstract public function cancel($watcherId);
 
--- a/src/Loop/Driver.php
+++ b/src/Loop/Driver.php
@@ -17,7 +17,7 @@ abstract class Driver
     /**
      * @var string
      */
-    private $driverId;
+    private $watcherIdPrefix;
 
     /**
      * @var string
@@ -34,7 +34,7 @@ abstract class Driver
      */
     public function __construct()
     {
-        $this->driverId = self::$nextDriverId++ . '-';
+        $this->watcherIdPrefix = self::$nextDriverId++ . '-';
     }
 
     /**
@@ -322,7 +322,7 @@ abstract class Driver
      */
     final protected function createWatcherId()
     {
-        return $this->driverId . $this->nextWatcherId++;
+        return $this->watcherIdPrefix . $this->nextWatcherId++;
     }
 
     /**
@@ -338,9 +338,9 @@ abstract class Driver
      *
      * @throws InvalidWatcherException
      */
-    final protected function validateWatcherId($watcherId, bool $mustThrow)
+    final protected function validateWatcherId($watcherId, $mustThrow)
     {
-        if (0 !== \strpos($watcherId, $this->driverId)) {
+        if (0 !== \strpos($watcherId, $this->watcherIdPrefix)) {
             // the watcher was created by a different driver
             throw new InvalidWatcherException($watcherId, 'A watcher was passed to the wrong driver.');
         }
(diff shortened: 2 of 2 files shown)


src/Loop.php
```diff
@@ -295,11 +295,14 @@ public static function disable($watcherId)
      * Cancel a watcher.
      *
      * This will detatch the event loop from all resources that are associated to the watcher. After this operation the
-     * watcher is permanently invalid. Calling this function MUST NOT fail, even if passed an invalid watcher.
+     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by this
+     * driver but is no longer valid.
```
joshdifabio: This block refers to "this driver", but this doesn't really make sense in the context of `Loop`. Not sure about this since currently `Loop` and `Driver` have the same doc blocks.

joshdifabio:
What do people think about adding a `WatcherIdentifierExpired` exception class (there might be a better word than expired) to distinguish between truly invalid watcher identifiers and ones which have simply already expired?

src/Loop.php
```diff
@@ -295,11 +295,14 @@ public static function disable($watcherId)
...
      *
      * @param string $watcherId The watcher identifier.
      *
      * @return void
+     *
+     * @throws InvalidWatcherException If the watcher identifier is invalid.
```
kelunik: Better: If the watcher identifier does not belong to this driver.

src/Loop/Driver.php
```diff
@@ -285,4 +303,56 @@
...
+
+    /**
+     * Create a watcher identifier.
+     *
+     * Any watcher identifiers returned to clients by the driver MUST be created using this method. The returned watcher
+     * identifier SHALL be unique for the duration of the PHP process.
+     *
+     * @return string The watcher identifier.
+     */
+    final protected function createWatcherIdentifier()
```
kelunik: I think we should call this `createWatcherId`, just like the properties.

joshdifabio: Makes sense.

src/Loop.php
joshdifabio: Good idea.

src/Loop/Driver.php
```diff
@@ -285,4 +303,56 @@
...
+     *
+     * @param string $watcherId The watcher identifier.
+     * @param bool $mustThrow Whether the method MUST throw. For example, if `enable()` was called with an unrecognised
+     * watcher identifier then this should be set to `true`, but if `cancel()` was called it should be set to `false`.
+     *
+     * @throws InvalidWatcherException
+     */
+    final protected function validateWatcherIdentifier($watcherId, bool $mustThrow)
+    {
+        if (!\preg_match('{^(?P<driver_id>[a-z]+)-(?P<watcher_id>[a-z]+)$}', $watcherId, $matches)) {
```
kelunik: Using regular expressions here is somehow a overkill. A simple `explode` does the job.

src/Loop.php
joshdifabio: Although it would also throw if the watcher identifier is truly invalid. E.g., wrong format, not yet issued.

src/Loop/Driver.php
joshdifabio: The `preg_match` simplifies the `strcmp` below since we then know that the string is in the correct format.

src/Loop.php
kelunik: We have to change the definition of invalid then.

joshdifabio: It seems to me that there are three possibilities: wrong driver, invalid (not actually issued by a driver), or an expired identifier.

src/Loop/Driver.php
bwoebi: Uh, can I please turn off this type of expensive verification on production?

joshdifabio: This function would only be called in error cases or when `cancel()` is called on a non-existent watcher. Is that still a problem? I guess the latter could be quite common?

bwoebi: It can at least.

src/Loop.php
kelunik: 4. bogus identifier

bwoebi: bogus = invalid in this context

src/Loop/Driver.php
joshdifabio: Doing some tests in the command line, `explode` takes virtually as long as `preg_match`, although we're talking about something like 200ms for 1m iterations. Perhaps we should simply check that the watcher has been passed to the correct driver using `0 == strpos($watcherId, "$driverId-")`, which is faster, and not make any other checks? This should rule out the important bug (wrong driver) without sacrificing performance?

src/Loop.php
joshdifabio: Yeah, invalid = bogus I think, as both simply mean that there is a definite programming error somewhere.

src/Loop/Driver.php
```diff
@@ -285,4 +303,56 @@
...
+     *
+     * Any watcher identifiers returned to clients by the driver MUST be created using this method. The returned watcher
+     * identifier SHALL be unique for the duration of the PHP process.
+     *
+     * @return string The watcher identifier.
+     */
+    final protected function createWatcherIdentifier()
+    {
+        if (!isset($this->driverId)) {
+            $this->driverId = self::$nextDriverId++;
```
trowski: This should go in a constructor.

joshdifabio: I would prefer that, too, but it adds something extra for Driver implementations to do. (I.e. Drivers MUST call parent::__construct() within their constructors.) I guess that's okay?

trowski: I see no problem with that. It's expected when you're extending a class with a constructor.

joshdifabio:
Unless I'm missing something, there isn't a significantly more efficient way of checking the format of a watcher than `preg_match()`, but I think the only important thing is erroring when a watcher identifier is passed to the wrong driver instance. Right? I don't think validating the format of the identifier is really very useful.

trowski:
Can we add to the spec that it is acceptable to not guarantee unique-per-process identifiers in production? I would definitely be a supporter here if loops can skip it if `zend.assertions = 0`.

joshdifabio:
> Can we add to the spec that it is acceptable to not guarantee unique-per-process identifiers in production?

What's the reasoning? Have you checked the latest version of the changes?

rdlowrey:
What if the spec instead allows loops to receive a callable (or `IdFactory`) responsible for generating watcher IDs? In that way you avoid the fcall overhead for folks who don't need it at the minimal cost of an additional `if` check (because increment operators could be used if the factory wasn't set). 

I'm concerned about the performance implications of adding a regex check on every watcher creation. It would seriously inhibit loop performance ...

bwoebi:
I think I can live with the current version of the PR.

kelunik (commented):
We can't use a bool type declaration.

joshdifabio:
> I'm concerned about the performance implications of adding a regex check on every watcher creation. It would seriously inhibit loop performance ...

There isn't a regex check on every watcher creation. Previously there was a regex check for unrecognised watcher identifiers but I've removed that. Please take a look at the latest PR and see if you still think the functionality needs to be opt-in.

> We can't use a bool type declaration.

Well spotted, thanks.

src/Loop/Driver.php
```diff
@@ -322,7 +323,7 @@ public function __construct()
      */
     final protected function createWatcherId()
     {
-        return $this->driverId . $this->nextWatcherId++;
+        return $this->nextWatcherId++;
```
bwoebi: https://3v4l.org/WJVEg Doesn't work :-/

joshdifabio: Aha. That's a shame. Thanks.

joshdifabio: I've changed it back.

joshdifabio:
I'll squash the commits if we decide to merge this.

joshdifabio:
@rdlowrey @trowski The overhead here is a single function call when creating a watcher, and two more (a method which in turn calls `strpos()`) when a loop driver encounters an unrecognised watcher. The benefit is that we can guarantee that clients shall never unknowingly enable/disable/cancel watchers from the wrong driver instance. Looking at the implementations of Amp's loops, for example, this overhead really seems very minimal. Do you still think this should be opt-in?

kelunik:
I'm not sure about the must throw boolean option. I think drivers should just call the method and throw afterwards if they're in enable or so.

joshdifabio:
> I'm not sure about the must throw boolean option. I think drivers should just call the method and throw afterwards if they're in enable or so.

Yes, I think I agree.

src/Loop/Driver.php
```diff
@@ -285,4 +311,35 @@
...
+     *
+     * @return string The watcher identifier.
+     */
+    final protected function createWatcherId()
+    {
+        return $this->watcherIdPrefix . $this->nextWatcherId++;
+    }
+
+    /**
+     * Validate a watcher identifier which was passed to the driver by a client.
```
kelunik: I'm not a native speaker, should the `was` instead be `has been`?

src/Loop/Driver.php
```diff
@@ -285,4 +311,35 @@
...
+     * expensive and is not typically necessary.
+     *
+     * @param string $watcherId The watcher identifier.
+     *
+     * @throws InvalidWatcherException If the watcher identifier does not belong to this driver.
+     */
+    final protected function validateWatcherId($watcherId)
+    {
+        if (0 !== \strpos($watcherId, $this->watcherIdPrefix)) {
+            throw new InvalidWatcherException($watcherId, 'A watcher was passed to the wrong driver.');
```
kelunik: The exception message should contain the `watcherId` here, so e.g. `"Watcher '{$watcherId}' has been passed to the wrong driver. Watchers passed to this driver must start with '{$this->watcherIdPrefix}'."`.

src/Loop/Driver.php
joshdifabio: I guess that as the invocation of the driver by the client with the invalid watcher has not yet returned, _has been_ would be better. I'll change it. Thanks.

src/Loop/Driver.php
joshdifabio: Sounds good.

joshdifabio: The only thing I'd say is that the terms _watcher_ and _watcher identifier_ are being used interchangeably. Perhaps we should reconcile that in the future.

src/Loop/Driver.php
```diff
@@ -285,4 +311,39 @@
      * @return null|object|resource The loop handle the event loop operates on. `null` if there is none.
      */
     abstract public function getHandle();
+
+    /**
+     * Create a watcher identifier.
+     *
+     * Any watcher identifiers returned to clients by the driver MUST be created using this method. The returned watcher
```
kelunik, line 318: I think `Every watcher identifier` is better.

kelunik: I think we can also remove the `returned to clients`, so we end up with `Every watcher identifier MUST be created using this method.`

src/Loop/Driver.php
```diff
@@ -285,4 +311,39 @@
...
+     *
+     * The driver MUST call this method when it is provided with a watcher identifier which it does not recognise. The
+     * driver SHOULD NOT call this method every time it receives a watcher identifier, as doing so would be relatively
+     * expensive and is not typically necessary.
+     *
+     * @param string $watcherId The watcher identifier.
+     *
+     * @throws InvalidWatcherException If the watcher identifier does not belong to this driver.
+     */
+    final protected function validateWatcherId($watcherId)
```
kelunik, line 339: Given that it doesn't return a boolean I'd rather call it `assertOwnWatcherId` or something like that.

src/Loop.php
```diff
@@ -295,11 +295,14 @@ public static function disable($watcherId)
      * Cancel a watcher.
      *
      * This will detatch the event loop from all resources that are associated to the watcher. After this operation the
-     * watcher is permanently invalid. Calling this function MUST NOT fail, even if passed an invalid watcher.
+     * watcher is permanently invalid. Calling this function MUST NOT fail if the passed watcher was issued by the
+     * current driver but is no longer valid.
```
kelunik, line 299: I think it's enough here to write: _"Calling this function MUST NOT fail if the passed watcher was issued by the current driver."_.

src/Loop/Driver.php
kelunik: @joshdifabio Could you do that change then?

bwoebi:
@joshdifabio Can you please look at @kelunik comments?

joshdifabio:
I'll aim to rebase and apply Niklas's recommendations this evening.

kelunik:
@joshdifabio ping.

kelunik:

kelunik:
Just for the record: I'm currently facing an issue where a `__destruct` cancels a watcher of another loop.

joshdifabio:
> Just for the record: I'm currently facing an issue where a `__destruct` cancels a watcher of another loop.

Scary. Make it impossible.

PR #138, opened 2017-01-09
Disallow other drivers when the Loop accessor has been used outside of Loop::execute

kelunik:
This protects applications creating watchers etc. outside of `Loop::execute` to run the right loop. I'm not sure, but it might create false positives.


kelunik:
Closing in favor of the way better #139.

PR #141, opened 2017-01-13
Define 'tick'

kelunik:


src/Loop/Driver.php
```diff
@@ -15,7 +15,14 @@
     private $registry = [];
 
     /**
-     * Start the event loop.
+     * Run the event loop.
+     *
+     * One iteration of the loop is called one "tick". A tick covers the following steps:
+     *
+     *  1. Activate watchers created / enabled in the last tick / before `run`.
```
bwoebi: With \`run\` you are referencing the method? Please suffix `()` then to make this more obvious.

src/Loop/Driver.php
```diff
@@ -15,7 +15,14 @@
...
+     *  2. Execute all enabled defer watchers.
+     *  3. Execute all due timers, each timer only once per tick.
+     *  4. Execute actionable stream watchers, each only once per tick.
```
bwoebi: Uh, there is no order guarantee between timers and stream watchers. Merge these two into a single step.

src/Loop/Driver.php
kelunik: Fixed.

src/Loop/Driver.php
kelunik: Fixed, once per tick should be fine?

bwoebi: yes, that's fine.

bwoebi (approved):

src/Loop/Driver.php
```diff
@@ -15,7 +15,13 @@
...
 
     /**
-     * Start the event loop.
+     * Run the event loop.
+     *
+     * One iteration of the loop is called one "tick". A tick covers the following steps:
+     *
+     *  1. Activate watchers created / enabled in the last tick / before `run()`.
+     *  2. Execute all enabled defer watchers.
+     *  3. Execute all due timer and actionable stream callbacks, each only once per tick.
```
trowski: Can signals be executed at any time or should they be included in this step? All current implementations execute pending signals only in this step, but should that be part of the spec?

trowski (changes requested):

src/Loop/Driver.php
bwoebi: Yes, they should. Missed that.

kelunik: Changed.

trowski (approved):

PR #145, opened 2017-01-29
Resolve #142 by adding a run() method

bwoebi:
As per: https://github.com/async-interop/event-loop/pull/142#issuecomment-275927251

Also adding a shutdown function at the end in order to ensure that the loop level will be correct (i.e. people able to use Loop::run() in a shutdown function).

\cc @davidwdan @trowski @mbonneau

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -89,14 +89,10 @@ final class Loop
      *
      * @return void
      *
-     * @see \AsyncInterop\Loop::run()
+     * @see \AsyncInterop\Loop\Driver::run()
      */
     public static function run()
     {
-        if (self::$level > 0) {
-            throw new \RuntimeException("The loop can only be run while not yet running.");
-        }
-
         $driver = self::$driver ?: self::get();
         self::$level++;
 
@@ -455,10 +451,3 @@ final class Loop
         // intentionally left blank
     }
 }
-
-// Reset the $level in order to be able to use ::run() inside a shutdown handler
-$f = function()
-{
-    self::$level = 0;
-};
-register_shutdown_function($f->bindTo(null, Loop::class));
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -43,7 +43,7 @@ final class Loop
     public static function setFactory(DriverFactory $factory = null)
     {
         if (self::$level > 0) {
-            throw new \RuntimeException("Setting a new factory while running isn't allowed!");
+            throw new \RuntimeException("Setting a new factory while running isn't allowed.");
         }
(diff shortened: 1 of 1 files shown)


src/Loop.php
```diff
@@ -430,3 +455,10 @@ private function __construct()
         // intentionally left blank
     }
 }
+
+// Reset the $level in order to be able to use ::run() inside a shutdown handler
+$f = function()
+{
+    self::$level = 0;
+};
+register_shutdown_function($f->bindTo(null, Loop::class));
```
kelunik: Hmm... this might not be the first shutdown function to be executed.

kelunik:
It seems like this doesn't actually solve anything. People can just use `Loop::execute(function () {}, Loop::get());` and get the same result?

src/Loop.php
```diff
@@ -430,3 +455,10 @@ private function __construct()
         // intentionally left blank
     }
 }
+
+// Reset the $level in order to be able to use ::run() inside a shutdown handler
+$f = function()
```
nikic: Isn't this polluting the variable scope?

bwoebi: given the composer autoloader, no.

src/Loop.php
bwoebi: I do not know of any way to prioritize that.

trowski: If this was done in a file included immediately by composer this would be much more likely… but still not guaranteed.

Any particular reason for requiring the level to be 0 to call `Loop::run()`? We don't have this restriction for execute, even when passing the loop driver returned from `Loop::get()`.

kelunik: Just not having `run` solves that, too.

bwoebi: No, it doesn't. you cannot use setFactory() there event though you're at top-level...

kelunik: @bwoebi (1) I don't think we need `setFactory()` in shutdown hooks and introduce hacks for it. (2) I'd like to allow `setFactory` everywhere by removing the `$driver = null` reset here.

joshdifabio: I'm not sure the limited reward justifies the added complexity here, especially since the invocation order of shutdown callbacks can't even be relied upon.

trowski: I pushed a commit dropping the level requirement… since shutdown function order can't be relied upon (and some apps even start the loop in a shutdown function) it just isn't worth trying to enforce it. Only so much we can do to protect users from themselves…

kelunik:
Can we close this one?

PR #147, opened 2017-01-30
Remove setFactory() side-effect and add resetDriver()

trowski:
Invoking `Loop::setFactory()` removes the active `Driver` instance as a side-effect. This PR removes this side-effect and requires a new `DriverFactory` instance be passed to `Loop::setFactory()` (instead of being nullable).

To remove the current `Driver` instance and force a new one to be created, a `Loop::resetDriver()` method is added. This method cannot be invoked while the loop is running.

### DIFF
--- a/src/Loop.php
+++ b/src/Loop.php
@@ -271,9 +271,8 @@ final class Loop
     /**
      * Enable a watcher.
      *
-     * Watchers (enabling or new watchers) MUST immediately be marked as enabled, but only be activated (i.e. callbacks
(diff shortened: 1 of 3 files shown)


kelunik:
Why did you remove the nullability?

trowski:
@kelunik The nullability seemed pointless… why would you explicitly remove the factory without setting another?

kelunik:
Because tests or whatever.

trowski:
Testing maybe… so I'll put it back.

kelunik:

PR #153, opened 2017-03-09
Remove Loop::execute scoping

### LINKED ISSUE #148: Scoping
We have quite a few PRs and the discussion is spread accross those PRs instead of being in one issue. I know, adding yet another discussion thread seems kind of pointless then, but I think it's important to gather some clear statements about expectations regarding scoping.

**PRs**

 - #139 
 - #142 
 - #144 
 - #145 
 - #149 

**My own statement**

I think a clear scope avoids bugs with different drivers. We allow passing a driver to be run to `Loop::execute` and should therefore only rely on the driver being accessible with the `Loop` accessor inside that `Loop::execute` scope.

It has been mentioned that this would prevent the adapter for React having a correct `run` method and additionally would prevent first setting things up and then running the loop.

The React adapter currently uses `Loop::execute(function () {}, Loop::get());` to run the default / current loop. This continues to work with a strict `Loop::execute` scope (#139) as long as the whole application is wrapped in `Loop::execute`. It uses a stacked driver then, but as both are the same, everything will work fine.

Another solution for the adapter is to use an explicit driver first and then simply pass it to `Loop::execute` as React uses an explicit loop that's passed around anyway.

```php
<?php

$loop = new LoopAdapter($interopLoop);
$connector = new DnsConnector($loop);
// ...
$loop->run(); // Executes Loop::execute
```

Registering events on a specific loop instance without the accessor will always register those events on the right driver (obviously) and is therefore totally fine outside of `Loop::execute`.

But after all, this is something which doesn't affect any library code. It only affects tests and application code, which both have to be changed anyway to make use of the accessor and interop loop.

I really want to avoid having `resetDriver` and a default driver at all if we can have a clear scoping and thus avoid bugs with different drivers and forgot-to-reset drivers in tests. It's going to be less confusing for newcomers if the following code just errors instead of just not working, because it's using two different loops.

```php
<?php

Loop::defer(function () { /* never executed */ });

// Uses a new driver and thus "ignores" the previous `Loop::defer()`
Loop::execute(function () { print "hello"; });
```

I'm very interested in opinions especially from @WyriHaximus @jsor @cboden @davidwdan @mbonneau

kelunik:
This allows using `Loop::run();` instead of putting the entire
application into a closure. A default loop can be setup using Composer's
`"files"` autoload feature. Scoping can be implemented on top by testing
frameworks or other libraries.

Closes #148.


src/Loop.php
```diff
@@ -59,50 +35,14 @@ public static function setFactory(DriverFactory $factory = null)
...
-
-        $driver = self::$factory->create();
-
-        if (!$driver instanceof Driver) {
-            $type = is_object($driver) ? "an instance of " . get_class($driver) : gettype($driver);
-            throw new \Exception("Loop driver factory returned {$type}, but must return an instance of Driver.");
-        }
-
-        return $driver;
+        $driver = self::$driver ?: self::get();
```
sagebind, line 44: What is the purpose of this `get()` call if it just returns the value of `$driver`?

kelunik:

